PR #1, opened 2018-07-07
some basic cleanup, dropping some axioms.

gmalecha:
This is just some basic cleanup.


Lysxia:
Nice! I was starting to think that axiom was necessary, it's cool that it's not!

Issue #6, opened 2018-07-13
Use ExtLib's monad type classes?


gmalecha:
Do people have opinions about pulling in ExtLib's typeclasses for monads?

bcpierce00:
I like the ExtLib monads, and we use them all the time internally.  (I’ll like them even better when they allow binding patterns — can they do that, yet?)

But there’s some feeling that, since there isn’t a community-standard monad library at the moment, we should try not to force people to use the one we prefer.  Would it be easy to use ExtLib internally but leave enough room for clients to make different choices?

   - B

gmalecha:
Clients can always add their own instances for other type classes if they wish to use another library. By using ExtLib here we would require them to install ExtLib. We could (with other library writers) separate the monads from ExtLib and depend specifically on them. If we can get buy-in from other authors, we can prevent fracturing.

There was a PR to support binding patterns, I need to release a new opam version with suppoet for it.

bcpierce00:
> Clients can always add their own instances for other type classes if they wish to use another library. By using ExtLib here we would require them to install ExtLib. We could (with other library writers) separate the monads from ExtLib and depend specifically on them. If we can get buy-in from other authors, we can prevent fracturing.

I think a dependency on ExtLib is not terrible — it’s just one more OPam install.  

Having said that, I think a separate monad library might get better uptake in the community (especially if it were structured to make it a little easier to find things — I always find the current organization a bit hard to navigate when I’m looking for things).
> There was a PR to support binding patterns, I need to release a new opam version with suppoet for it.
> 
> 

Great!

   - B

gmalecha:
@bcpierce00 It would be great to hear any suggestions that you have about the ExtLib monads, especially if we split it into another library.

bcpierce00:
Just simplifying the directory structure would go a long way!  Also adding a tiny bit of documentation — just an index of what’s there in the README file would be a big help.

gmalecha:
This is done.

PR #9, opened 2018-07-14
`mfix` needs to give the monad structure to the body.

gmalecha:
When trying to use the definition of `mfix`, I noticed that I forgot to give the body access to a monad instance for `m`, which prevents it from using both `inj` and `rec`.

I am using the `ExtLib` definition of `Monad`, though in #6 we decided that it might be better to pull out a separate library. We can do that before merging this if we want.

### DIFF
--- a/theories/ITree.v
+++ b/theories/ITree.v
@@ -1,3 +1,5 @@
+Require Import ExtLib.Structures.Monad.
+
 Set Implicit Arguments.
 Set Contextual Implicit.
 
@@ -77,21 +79,19 @@ Qed.
 
 End Core.
 
-Definition liftE (E : Type -> Type) (X : Type)
-           (e : E X) : itree E X :=
-  Vis e Ret.
-
-(** Monadic return *)
-Definition ret : forall {E R}, R -> itree E R := @Ret.
-
 (** Monadic bind *)
 (* We insert a [Tau] in the [Ret] case to make programs/specifications
    neater. This makes [itree] no longer a monad structurally,
    but it remains one in a looser sense as long as [Tau] is
    interpreted as the identity. *)
-Definition bind {E R S}
+Definition bind' {E R S}
            (t : itree E R) (k : R -> itree E S) : itree E S :=
   Core.bind t (fun r => Tau (k r)).
+
+Global Instance Monad_itree {E} : Monad (itree E) :=
+{ ret := @Ret E
+; bind := @bind' E
+}.
 
 (* A lemma to unfold [bind]. *)
 Lemma bind_def E R S :
@@ -99,11 +99,16 @@ Lemma bind_def E R S :
     bind t k =
     Core.bind_body t (fun t' => bind t' k) (fun r => Tau (k r)).
 Proof.
-  unfold bind.
+  simpl bind; unfold bind'.
   intros s k.
   rewrite Core.bind_def_core.
-  auto.
+  reflexivity.
 Qed.
+
+
+Definition liftE (E : Type -> Type) (X : Type)
+           (e : E X) : itree E X :=
+  Vis e Ret.
 
 Definition sequ {E R S}
            (t : itree E R) (u : itree E S) : itree E S :=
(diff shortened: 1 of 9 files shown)


theories/ITree.v
```diff
@@ -77,34 +79,37 @@ Qed.
...
-
-(** Monadic return *)
-Definition ret : forall {E R}, R -> itree E R := @Ret.
-
 (** Monadic bind *)
 (* We insert a [Tau] in the [Ret] case to make programs/specifications
    neater. This makes [itree] no longer a monad structurally,
    but it remains one in a looser sense as long as [Tau] is
    interpreted as the identity. *)
-Definition bind {E R S}
```
gmalecha: I actually don't understand the reason for adding the extra `Tau` here.

theories/ITree.v
```diff
@@ -178,23 +183,3 @@ Definition hoist {E F R}
 Definition when {E}
            (b : bool) (body : itree E unit) : itree E unit :=
   if b then body else ret tt.
-
-Module ItreeNotations.
```
gmalecha: We inherit the notation from `ExtLib`.

Lysxia:
On why we add `Tau` in `bind'` (replying to the comment here for better visibility). Without `Tau`, a simple loop `cofix loop := body;; loop` simplifies in one step to

```
cofix loop :=
  match body with
  | Ret _ => loop
  | ...
  end
```

Even assuming this is productive, for many reasonable definitions of `body` (if `body` contains loops, or starts with a conditional), this fails the guardedness condition. (Basically, guardedness checking succeeds only when the definition of `body` is exclusively made of constructors, with no `cofix` or `match`). Replacing the inner `loop` with `Tau loop` works around that.

gmalecha:
@Lysxia That makes sense, it isn't necessary if you use `mfix` though. If you define your loops using `mfix`, you will always get `Tau`s when you need them, i.e. when the body of the loop is not guarded by anything else.

Lysxia:
That's a fair point. In fact `mfix` may be a more general solution.

This `bind'` which inserts `Tau` is only a partial fix for productivity problems. There is another problem when `loop` is not tail-called. For example in `cofix loop := x <- body;; xs <- loop;; cont x xs`, the second bind expands to an undesirable `match loop with ...`. I currently avoid that with a bit of CPS (currently by hand) but it's not a complete fix either (the problem in my previous comment actually comes back). Another more difficult pattern for this coinductive type is recursive calls that occur within loops, which happens frequently in tree/graph traversals.

`mfix` seems to sidestep all those issues. I'm not sure what the downsides are. Is it a lot of syntactic overhead to define small loops with it compared to `cofix` (when that works)?

gmalecha:
Currently it isn't as nice to use as I would like it to be, but I think that can be mitigated to some extent by a combination of specializations for particular patterns (e.g. your `spin`, `forever`, a non-dependent fixpoint, and mutual fixpoints) and notations. I think that there will always be some overhead, but it might be made up for by the proof rules that you get for `mfix` which aren't even possible to write for `cofix`.

In its current form, the unfolding lemma should give you exactly what you want by beta-reduction so nothing special needed there.

gmalecha:
While looking over my previous notes, I noticed that another option (which doesn't require `Monad`) is to make `mfix_body` have the following meaning

```coq
Definition mfix_body (eff : Type -> Type) (dom : Type) (codom : dom -> Type) : Type :=
  forall eff' (inj : forall t, itree eff t -> itree eff' t) (rec : forall x : dom, itree eff' (codom x)), forall x : dom, itree eff' (codom x).
```

This might be easier to work with.

gmalecha:
With the latest change, we can once again drop the dependence on `ExtLib`.

gmalecha:
@Lysxia @Zdancewic @Chobbes @bcpierce00 Any opposition to merging this? I could remove the ExtLib monad dependencies, but I would still need monoid for the `eff_hom_w` (i.e. the writer monad) instance.

I've actually written some semantics on top of these definitions, and they seem to work out reasonably well. I will open another pull request with a semantics for Imp based on these definitions.

PR #10, opened 2018-07-15
simplify the definition of bind to avoid unnecessary taus

gmalecha:
This simplifies the definition of `bind` based on the discussion in #9 . I haven't had a chance to update the proofs in `Equivalence.v` yet. Are you guys building on top of those proofs?

### DIFF
--- a/CONVENTIONS.md
+++ b/CONVENTIONS.md
@@ -1,0 +1,14 @@
+# Coq Conventions
+
+## Naming conventions:
+
+- [python_case] for definitions
+- [CamelCase] for constructors and classes
+
+Variables:
+- [E E1 E2 F F1 F2 : Type -> Type] effect types
+- [X Y Z : Type] effect output types
+- [e : E X] effects
+- [R S : Type] return/result types ([Ret : R -> itree E R])
+- [t u : itree E R] itrees
+- [k h : R -> itree E S] itree continuations (iforests?)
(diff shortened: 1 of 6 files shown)


Lysxia:
We aren't yet.

gmalecha:
Note that with this PR, the `Equivalence.v` file no longer compiles.

gmalecha:
I also just realized that this means that `hoist` should no longer be necessary since `forall i f, hoist f i = hom (fun _ e => Vis e Ret) i` since `bind` doesn't introduce extra `Tau`s now.

I believe it is that case (though it would be difficult to state) that there are only two ways to get `Tau`s if you use regular combinators (i.e. `bind`, `ret`, `mfix`): 1. You write a program that is equivalent to `spin`; 2. you use `hom` to interpret an effect "fully", i.e. the meaning of an effect is a `ret`.

gmalecha:
I did a bit more refactoring trying to split out some of the functionality (I'm a fan of small, self-contained files). In going through `Effects.v`, it seems like there are a lot of places where we re-implement `hom` (I renamed it to `interp`). Now that we don't need to worry about "economical" mappings (since `bind` doesn't introduce `Tau`s), we should be able to consolidate these. In general, for each new effect we have, we should be able to pick an appropriate monad to interpret it and then use the appropriate `eff_hom`. For example, if we pick the identity monad, we use `eff_hom`, if we use the state monad we pick `eff_hom_s`. There can be similar `eff_hom`s for other monads, one of the more exotic ones is `interp_prop` which interprets into the monad formed by `fun x => x -> Prop` (with the usual `bind` and `ret`).

PR #14, opened 2018-07-26
[Don't merge] Discussion on related project

vukovinski:
I recently finished a first sprint implementation of a small embedded language that models interaction with the API of a InterPlanetary File System node - [fsharp-ipfs-dsl](https://github.com/cloveekprojeqt/fsharp-ipfs-dsl/).

The current implementation has tens of sub-DSL's that are joined in the sum type IpfsDSL, and a Free Monad construction over the IpfsDSL provides the semantics needed to implement interpreters for the embedded language. However, this does not satisfy me, as it precludes the ability for more efficient interaction, more specifically, I would like to have at least some notion of concurrency and elimination between the various sub-DSL instructions and within them.

After reading your work, it seems to me that a visible effects system combined with a freer monad (~= itree?) is the construction I possibly need to satisfy these requirements succinctly. 

I don't want to waste your time, having found your project via GitHub recommendations, but I would like to write here in this thread the results of my adventures as I find it a great opportunity for learning. Feel free to add suggestions.

Thanks

### DIFF
--- a/theories/ITree.v
+++ b/theories/ITree.v
@@ -10,6 +10,8 @@ Set Contextual Implicit.
     [R] and every node is either a [Tau] node with one child, or a
     branching node [Vis] with a visible effect [E X] that branches
     on the values of [X]. *)
+
+(** This is almost exactly the kind of more powerful free monad I needed *)
 CoInductive itree (E : Type -> Type) (R : Type) :=
 | Ret (r : R)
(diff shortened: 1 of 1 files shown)


Lysxia:
We definitely welcome discussion on this topic! This is a bit of an unusual format but people can unsubscribe from this thread if they want.

"Interaction tree" is indeed what we name the freer monad in this project.

Lysxia:
(Closing old PRs)

PR #17, opened 2018-08-29
Finish [untaus_equiv].

lastland:
I was looking at the code when I saw an admitted lemma, so I tested my understanding of the code by finishing the proof. Hopefully, this does not conflict with other people's ongoing work...


Lysxia:
Looks good to me!

PR #20, opened 2018-09-21
Alternative representation of effects

Lysxia:
This is still very experimental, and whether the change is really beneficial remains to be determined.

Instead of an indexed type `Type -> Type` an effect type is now a dependent record `{ action : Type; reaction : action -> Type }` and the `Vis` constructor now has type...

```coq
Vis : forall (e : action) (k : reaction e -> itree), itree
(* as opposed to: forall (X : Type) (e : E X) (k : X -> itree), itree *)
```

The point is that `Vis` has one fewer index, and that equality between effect constructors `e` can be expressed homogeneously `e = e'`, as opposed to the heterogeneous `existT _ X (e : E X) = existT _ X' (e' : E X')` that currently get generated by some uses of `inversion` with predicates about itrees. (But we still get heterogeneous equalities between continuations `existT _ e (k : reaction e -> _) = existT _ e' (k' : reaction e' -> _)`.)

There is a bit of type inference that needs to be figured out: in `Imp.v` I had to specialize some occurrences of `do` (function that lifts raw effects to trees: `(E -< F) -> E X -> itree F X` (old) or `(E -< F) -> forall e : E, itree F (reaction e)` (new)).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,14 @@ install: Makefile.coq all
 uninstall: Makefile.coq
 	$(MAKE) -f $< $@
 
+examples: example-imp example-lc
+
+example-imp: examples/Imp.v
+	coqc -Q theories/ ITree examples/Imp.v
+
+example-lc: examples/stlc.v
+	coqc -Q theories/ ITree examples/stlc.v
+
 Makefile.coq: _CoqProject
 	coq_makefile -f _CoqProject -o Makefile.coq
 
--- a/examples/Imp.v
+++ b/examples/Imp.v
@@ -7,10 +7,10 @@ Require Import ExtLib.Structures.Monad.
 Require Import ExtLib.Structures.Traversable.
 Require Import ExtLib.Data.List.
 
-Require ITree.ITree.
+Require Import ITree.ITree.
 Require Import ITree.Morphisms.
 Require Import ITree.Fix.
-Require Import ITree.Effect.
+Require Import ITree.OpenSum.
 
 Import MonadNotation.
 Local Open Scope monad_scope.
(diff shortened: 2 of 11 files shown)


gmalecha:
I like this because it makes the universes non-expansive which makes it possible to implement `join`. Is there anything that we can't do with this representation that we would like to do?

Lysxia:
Is it not already possible to implement `join`? Or does something break at some point

```
Definition join {m} `{Monad m} {a} (t : m (m a)) : m a :=
                                  bind t (fun x => x).
```

gmalecha:
The universe level of itree is one larger than the level of it's argument, so I think it requires universe polymorphism.

Lysxia:
Any objections to merge this?

Ideally there should be a comparison of this style with the current one. I would hope that with most concrete `E : Type -> Type`, there is a way to convert equality of `existT E X (e1 : E X) = existT E X (e2 : E X)` to `e1 = e2`, and then use that to convert `existT _ X x = existT _ X y` to `x = y`.

gmalecha:
You mentioned that there was a little bit more work to do related to inference. Is that resolved? If so, I think we should merge it.

Lysxia:
Oh that's right I will take a look at that issue.

Lysxia:
One downside of such dependent types is that sometimes coq doesn't reduce as often as we would like it to... Here:

https://github.com/DeepSpec/InteractionTrees/blob/909a16b2e121c63db9a2656cab9eff7738bc0543/examples/IO.v#L17-L19

`liftE (Write n)` has type `itree _ (reaction (Write n))`, which can reduce to `itree _ unit`. However Coq doesn't want to reduce this (unless you force it with an annotation) so it remains as `itree _ (reaction (Write n))`, until it escapes the scope of `n`, when typechecking the `bind`.

Issue #23, opened 2018-10-29
reverting to the old definition of itree


aa755:
At BedrockSystems, I have defined some generic theories that (IMO and somewhat formally) provide a more general and succinct way to solve the problems which the new definition of `itree` seems to solve. 
By old definition, I mean the following where effects were just specified as `Type->Type`:
````
CoInductive itree (E : Type -> Type) (R : Type) :=
| Ret (r : R)
| Vis {X : Type} (e : E X) (k : X -> itree E R)
| Tau (t : itree E R).
````

Firstly, the old type is more succinct IMO.
Also, I think that @gmalecha  said that the new definition of `itree` makes it hard to use typeclass inference and needs canonical structures; I don't understand this aspect.

Based on my discussion with @gmalecha, I guess the main problem was equating the constituents of two `Vis` terms.
For example, previously, to equate `(Vis (e1:(E X1)) k1)`  and `(Vis (e2:(E X2)) k2)` one would have to say `{p:X1=X2 | (rew p in e1)=e2 /\ (rew [fun U => U -> itree E T] p in k1)=k2}`, which may look a bit ugly. In the new definition of `itree`, it has one less conjunct but it doesn't fundamentally look any simpler:
`{p:e1=e2 |  (rew [fun U => U -> itree E T] (f_equal reaction p) in k1)=k2}`.
Rewriting is still needed to equate the continuations (`k1, k2`). This rewriting won't simplify unless the proof `p` is `eq_refl`. 

For example, the following was and is STILL unprovable without using the UIP axiom 
`forall ... e k1 k2,  (Vis e k1) = (Vis e k2) -> k1=k2.`

In the new definition, you can avoid the UIP axiom if you require that the 
 `action` Type in `Effect` has decidable equality. 
I have a solution that is more general and does not require forgoing the succinctness of the old definition of `itree`: I only require that the set of return types of the actions be isomorphic to a type with decidable equality. The theory can be found in this file in my fork:
https://github.com/aa755/InteractionTrees/blob/ETypeType/theories/inversion.v
The main lemma, `invertTauEqGen`, produces an instance of
````
Class InvertTauEq (E:Type->Type):= invertTauEq:  forall {T X : Type} (e : E X)
                                                  (k1 k2 : X -> itree E T), 
      (Vis e k1) = (Vis e k2) -> k1=k2.
````
from an instance of

````
Class DecidableEffRet {E:Type->Type} {Code:Type}:=
  {
[...]

PR #24, opened 2018-10-29
inversion, coinduction using numbers, eutt complementary alternative

aa755:
This PR adds the following 3 files listed in order of dependencies (every file depends on the previously mentioned files). These files assume the old representation of `itree`.

1) inversion.v : inversion principle for inverting (Tau e k1=Tau e k2). This is explained in the issue https://github.com/DeepSpec/InteractionTrees/issues/23

2) natCoind.v:   coinduction using induction over natural numbers and continuity. The main advantage is that with this definition, coinductive proofs can be done by induction over natural numbers. There is no productivity-check problem. Also, unlike in paco, properties previously proved about the "coinductive" relations are easily available to use in later proofs.
The main lemma is [`equivCoind`](https://github.com/aa755/InteractionTrees/blob/1b9087e7156ffcad5d6ebb94236995d8af12c29f/theories/natCoind.v#L37).
 It is applied to `eutt` in the lemma [`equivCoindEutt`](https://github.com/aa755/InteractionTrees/blob/1b9087e7156ffcad5d6ebb94236995d8af12c29f/theories/natCoind.v#L93)

3) euttC.v : this defines `euttIndC`, an equivalent complementary alternative to `eutt`. Although `eutt` is easier to consume as a hypothesis, in many applications, I found `euttIndC` easier to prove as a conclusion. The main advantage is that when proving `euttIndC`, the "coinduction hypothesis" can be invoked just after peeling off a `Tau` on both the left and the right terms. An orthogonal advantage is that `euttIndC` is defined using induction over naturals using the machinery in the previous file. I have proved that `euttIndC` implues `eutt`. I think the other direction should be easier.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,10 @@ git clone https://github.com/DeepSpec/InteractionTrees.git && cd InteractionTree
 
 ```
 opam install coq-paco coq-ext-lib
+git clone https://github.com/aa755/SquiggleEq
+cd SquiggleEq
+make -j5
+make install
 ```
 
 Now you can build the project with:
--- a/_CoqConfig
+++ b/_CoqConfig
@@ -8,3 +8,6 @@ theories/Fix.v
 theories/UpTo.v
 theories/MutFix.v
 theories/Trace.v
+theories/inversion.v
+theories/natCoind.v
+theories/euttC.v
--- a/theories/euttC.v
+++ b/theories/euttC.v
@@ -1,0 +1,624 @@
+Require Import ITree.inversion.
+Require Import ITree.natCoind.
+Require Import Coq.Relations.Relation_Definitions.
+Require Import ITree.ITree.
+Require Import ITree.Equivalence.
+Require Import SquiggleEq.UsefulTypes.
+Require Import SquiggleEq.tactics.
+Require Import SquiggleEq.LibTactics.
+Require Import Coq.Classes.RelationClasses.
+(*
+Require Import Coq.Logic.FunctionalExtensionality.
+Import EqNotations.
+Require Import Coq.Logic.Eqdep_dec.
+Require Import EqdepFacts.
+ *)
+
+Section euttC.
+  Context {E: Type -> Type} {R:Type}.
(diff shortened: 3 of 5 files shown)


Lysxia:
At a glance, `natCoind` definitely seems nice to work with! I wonder what the paco developers think of this; surely they must have considered it. Have you encountered any issues with having to apply a hypothesis in that form to a concrete `nat` upfront?

As for eutt, the idea of also being able to match one `Tau` on both sides in one coinductive step is also what we've been using in our webserver project (modulo using Coq's native `CoInductive` instead of `natCoind` or `paco`). I chose to experiment with the current definition in InteractionTrees because the proof terms of `eutt t1 t2` are much more structured (as you've noticed), to see whether anything good would come out of it. In particular, in the proof of transitivity, the "coinductive layers" match nicely, without silly cases like having `Eutt_TauL` on both sides. It's true that so far it's quite a pain to work with because there are a lot more moving pieces, but I am still hoping to come up with more lemmas and tactics to help streamline the proofs.

Showing equivalence between the two `eutt` would be an interesting exercise, which seems quite hard because of the different ways they account for delays (as you noted in comments also).

aa755:
Continuity of functions on relations is not always provable. 
I have been unable to prove it for some functions. I have read papers hinting unprovability in some cases. But I have never seen a formal argument of unprovability for any example. 

In contrast, paco can be used even with non-continuous but monotone functions on relations. 

Sometimes, it is tricky to pick a particular `nat`, e.g. in the proof that `euttIndC` implies `eutt`. But I think analogous problems would show up in that proof even if done using paco,

gmalecha:
It seems like we've tamed the co-inductive proofs. I'm wondering if we should close this.

Zdancewic:
I agree that we should close this.  We may want to eventually revisit the approximation-based reasoning, but things have changed enough that doing so would probably warrant a new issue / pull request.

gmalecha:
@Zdancewic @Grain

Issue #25, opened 2018-10-31
Positive coinductive types are going to be deprecated


Lysxia:
For a description of the problem: https://github.com/coq/coq/pull/8670/files

We'll have to change `itree` to be a record type, with the variants described by a separate nonrecursive type, that I just happened to define to do some category theory in `Machine.v`.

gmalecha:
I have code for this in the other repository.

https://github.com/gmalecha/coq-interaction-trees/blob/master/theories/Eff.v

It shouldn't affect too many things, but probably best to do earlier rather than later

gmalecha:
Closed by #26

PR #26, opened 2018-10-31
Converting `itree` to a record

gmalecha:
This is meant to address #25.

**note this doesn't completely compile yet**

- porting the definitions isn't too difficult, porting the proofs is
  more complex. it would be nice if we could make this change in a
  fairly straight-forward way, but the extra layer of indirection gets
  in the way of reduction.


theories/ITree.v
```diff
@@ -5,35 +5,65 @@ From ITree Require Export Effect.
 
 Set Implicit Arguments.
 Set Contextual Implicit.
+(* note(gmm): this seems to prevent us from doing `destruct t`
```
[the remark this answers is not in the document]
gmalecha: Ah, I see. It needs to be a much more invasive change then, including changing the definitions of eutt and even eq for itree.

I can keep working on it.

gmalecha:
I haven't finished looking into it in great detail, but there might be some issues with the way that we define `eutt_0`.

Lysxia:
I think I see what you mean. `Ret` and `Vis` are no longer constructors, we now have to match on `RetF` and `VisF`, but they are interleaved with `do`s, which makes things rather complicated. Perhaps we need a generic way of lifting relations through `do`s.

gmalecha:
Yes. We essentially need a constructor like do that will apply an observe.

gmalecha:
I gave this a shot and it worked out ok. It still seems like there is a lot of annoying pieces to it. I'm interested to know if integrating with paco addresses some of these problems.

gmalecha:
Thanks! What's the story on this now? Is it nearing mergability?

Lysxia:
There still seems to be a long way to go before this stabilizes.

Having these `observe` everywhere still feels rather clunky, equations are now of the form `observe t = TauF t'`, forcing you to unfold and rewrite explicitly, where previously just a `subst` did everything because you had `t = Tau t'`. Using more notations instead of definitions might alleviate this but I'm not sure how well that scales up.

There has to be a good way of working with this kind of representation. I'd like to find some existing and related Agda or Coq code to draw inspiration from.

Lysxia:
I'm starting to get the hang of this.

One thing to know is that `cofix` now reduces under projections instead of `match` (which are now disallowed), but somehow `simpl` will not perform that reduction, `cbn` will. Note also the `Arguments` command for `bind_match` to force its unfolding (whereas previously it unfolded on its own).

gmalecha:
@Lysxia Does that mean that you're comfortable merging this? What needs to happen before we are ready for that?

Lysxia:
The main remaining task is to finish porting the proofs in Eq.UpToTaus (previously Equivalence), and then synchronize with projects currently using Interaction Trees (@Zdancewic @aa755 ?). Since this is a pretty big change, we should test this out before merging.

gmalecha:
@Lysxia @Zdancewic What is the status of the definitions and proofs in `Eq.UpToTaus`? Some of the definitions in that file seem a bit awkward (there might be a good reason for the particular phrasing, I'm not sure). I'm wondering if it is necessary to hold block this PR on fixing those proofs given the exploration that is currently happening wrt paco.

Lysxia:
I agree that exploration has blocked this PR long enough. I'll admit the missing bits for now so this can be merged.

Lysxia:
Even though it still needs polishing I think we can merge this to move things forward and start building things on top of the new definition.

Lysxia:
Bikeshedding on notation.

It also seems like a bad idea to use `~` for `eutt` since the symbol is already used for `not`, though as a prefix operator. Should it use the same symbol as `eq_itree`?

gmalecha:
@Lysxia Notation is generally difficult. I think I was using `\approx` with unicode for a little while which isn't too bad, but can be annoying to type sometimes. I'll mostly defer to other people's opinions, but personally, I don't mind just not using a lot of notation.

theories/Eq/UpToTaus.v
```diff
@@ -547,110 +547,29 @@ Proof.
...
 Qed.
 
-(*
-Lemma finite_taus_bind_unalltaus {E R S}
-      (t t' : itree E R) (k : R -> itree E S) :
-  untaus t' t ->
-  finite_taus (t >>= k) <-> finite_taus (t' >>= k).
+(* [eutt] is a congruence wrt. [bind] *)
+Lemma eutt_bind {E R S} (t1 t2 : itree E R) (k1 k2 : R -> itree E S) :
+    t1 ~ t2 -> (forall x, k1 x ~ k2 x) -> (t1 >>= k1) ~ (t2 >>= k2).
```
gmalecha: We should really write this as a `Proper`, i.e. `Proper (itree_eq ==> pointwise_relation itree_eq ~ ==> itree_eq) bind`. That should make it easier to plug into type class resolution for rewriting etc.

theories/Eq/UpToTaus.v
```diff
@@ -547,110 +547,29 @@ Proof.
...
-  - reflexivity.
-Qed.
+Admitted.
 
-Lemma finite_taus_bind {E R S}
-      (t1 t2 : itree E R) (k1 k2 : R -> itree E S) :
-    t1 ~ t2 -> (forall x, k1 x ~ k2 x) ->
-    finite_taus (t1 >>= k1) -> finite_taus (t2 >>= k2).
+Lemma eutt_map {E R S} (f : R -> S) (t1 t2 : itree E R) :
+  t1 ~ t2 -> map f t1 ~ map f t2.
```
gmalecha: Same, we should write this using `Proper`. `Proper (eq ==> itree_equiv ==> itree_equiv) map` (maybe also generalizing over `eq` for the function).

theories/Eq/UpToTaus.v
```diff
@@ -547,110 +547,29 @@ Proof.
...
 Proof.
-  rewrite match_bind; now destruct t.
-Qed.
+Admitted.
 
-Lemma unalltaus_bind {E R S} (t : itree E R) (k : R -> itree E S) tk' :
-  unalltaus (t >>= k) tk' ->
-  (exists t', unalltaus t t' /\ unalltaus (t' >>= k) tk').
+Lemma eutt_forever {E R S} (t1 t2 : itree E R) :
+  t1 ~ t2 -> @forever _ _ S t1 ~ forever t2.
```
gmalecha: `Proper (eq_itree ==> eq_itree) forever`

theories/Eq/UpToTaus.v
```diff
@@ -547,110 +547,29 @@ Proof.
...
-      intro x. right. apply eutt_bind; auto.
-      destruct (H x) as [ | []]; auto.
-Qed.
-
-(* Failed naive attempt *)
-Lemma eutt_bind_bind_fail {E R S T}
-      (t : itree E R) (k : R -> itree E S) (h : S -> itree E T) :
-    (t >>= k >>= h) ~ (t >>= fun r => k r >>= h)%itree.
+Lemma eutt_when {E} (b : bool) (t1 t2 : itree E unit) :
+  t1 ~ t2 -> when b t1 ~ when b t2.
```
gmalecha: `Proper (eq_itree ==> eq_itree) (when b)`

Lysxia:
I've added the `Proper` instances, thanks for the help! I was only familiar with the heavyweight `Add Parametric Morphism` syntax.

gmalecha:
Let's merge?

Lysxia:
Good!

PR #27, opened 2018-10-31
Revert "Alternative representation of effects"

gmalecha:
This reverts commit 0f2e963e38dfe2e9ad359baad6272587fed2efcc.

Addresses #23


gmalecha:
I didn't update the IxEffect.v file, so this currently breaks things. I'll update that shortly.

PR #32, opened 2018-11-09
Using W-types to express effects that take interaction trees

gmalecha:
The definition works out fine, the problem is that you have to justify guardedness of your effect homomorphisms and I haven't found (though I haven't looked very hard) a good way to do this yet.

### DIFF
--- a/theories/REffect.v
+++ b/theories/REffect.v
@@ -1,0 +1,53 @@
+Require Import ExtLib.Data.HList.
+
+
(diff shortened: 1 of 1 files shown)


Lysxia:
Closing old issues/PR.

I note that we now have an `Events/Concurrency.v` module that demonstrates a similar construction, while reusing itree. But an ad-hoc `interp` would still have to be redefined to handle it.

PR #36, opened 2018-12-03
change Tau to notation?

Zdancewic:
The current master branch doesn't work cleanly with extraction because the definition of `Tau`, which is the body of `spin`, isn't allowed by OCaml (it doesn't allow an application on the right-hand-side of `let rec`).

This pull requests attempts to fix the issue by turning the `Tau` helper function into notation (which does extract cleanly). 

There are a couple questions:
- Is this the right solution?  
- Given that is , what is the right precedence for [Tau]
- Would we want to treat either [Ret] or [Bind] as notation instead of definitions?

### DIFF
--- a/examples/IO.v
+++ b/examples/IO.v
@@ -12,12 +12,12 @@ Definition example : itree IO unit :=
 Definition SOME_NUMBER := 13.
 
 Definition test_interp : itree IO unit -> bool := fun t =>
-  match t with
-  | Vis e k =>
+  match t.(observe) with
+  | VisF e k =>
     match e in IO X return (X -> _) -> _ with
     | Read => fun id =>
-      match k (id SOME_NUMBER) with
-      | Vis (Write n) _ => n =? SOME_NUMBER
+      match (k (id SOME_NUMBER)).(observe) with
+      | VisF (Write n) _ => n =? SOME_NUMBER
       | _ => false
       end
     | _ => fun _ => false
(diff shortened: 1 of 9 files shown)


Lysxia:
That looks alright. The only difference seems to be that whenever `Tau` is not applied, you now have to write `fun t => Tau t`, which isn't too bad. I'm slightly in favor of turning `Ret` and `Vis` into notations too, if only for consistency. They would also have the laziness properties during extraction that are closest to the previous version.

You can also define notations like this:

```coq
Notation Tau t := (do (TauF t)).  (* No quotes, no levels. *)
```

gmalecha:
I'm in favor of the change. I think for consistency, we should make notations for `Ret` and `Vis` as well. After that, we should merge.

Zdancewic:
OK -- I'll put this together now.

Zdancewic:
OK -- I changed Tau and Ret to be notations too.  The only slight drawback, as Li-yao mentioned, is the need to eta expand in a few places.  In the current code, mostly this is `fun x => Ret x`. 
Almost always it seems to occur in the pattern `Vis e (fun x => Ret x)`, which suggests that we add a definition (maybe even a coercion??) that lifts effects into itrees, something like:

```
Definition Eff e := Vis e (fun x => Ret x).
```

I'll merge this now.

Lysxia:
We have this function `liftE`.

Zdancewic:
Cool -- I'll push am update that uses liftE where appropriate.

Issue #34, opened 2018-11-25
Separate some of the core effects?


gmalecha:
Whenever you import `Effects`, you get a lot of effects. This can be convenient in some instances, but can also be very painful, for example if someone defines their own collection of effects and wants to use the name `Get`.

I'd like to propose separating some of the effects so that it is possible to `Require Import` just the sum effect (which there is no reason to ever overload) and get the definition of `do`.

What do you guys think?

bcpierce00:
Sounds good to me.

Lysxia:
The current situation for this issue is that the sum effect was split out (which I think is the most important bit), but the "standard" set of effects still needs reorganization.

Lysxia:
Fixed in #106

PR #39, opened 2018-12-11
Refactoring

Lysxia:
- Split out new module `StdEffects` from `OpenSums` (#34)
- Rename `ITree.ITree` to `ITree.Core` (core definitions), freeing up `ITree.ITree` for the standard way to open the library.
    + Operations defined in `Core` are further wrapped in an `ITree` module, the intent is that they should be used qualified. (I've had problems caused by shadowing `bind` and `map`.)
    + `itree` is still at the root of `Core` so it doesn't need to be qualified, but its actual name is now `ITree.Core.itree`; we could add notations inside the inner `ITree` module to "rename" it.
- Add `embed`, yet one more variant of `lift`, etc., but probably the easiest one to use when defining new effects (see `examples/IO.v` for example).


theories/Core.v
```diff
@@ -0,0 +1,232 @@
...
+ * using the monad laws up to a setoid.
+ * this would be *really* useful to a lot of projects.
+ *
+ * this will be especially important for this project because coinductives
+ * don't simplify very nicely.
+ *)
+
+(** Functorial map ([fmap]) *)
+(*
+Definition map_match {E R S}
```
gmalecha: Maybe drop this code?

theories/Core.v
```diff
@@ -0,0 +1,232 @@
...
+(* Instead of [pure := @Ret E], [ret := @Ret E], we eta-expand
+   [pure] and [ret] to make the extracted code respect OCaml's
+   value restriction. *)
+Instance Applicative_itree {E} : Applicative (itree E) :=
+{ pure _ := fun x => Ret x
+; ap _ _ f x := ITree.bind f (fun f => ITree.bind x (fun x => Ret (f x)))
+}.
+
+Global Instance Monad_itree {E} : Monad (itree E) :=
+{| Monad.ret _ := fun x => Ret x
```
gmalecha: Why the two different notations here?

theories/Core.v
Lysxia: As in drop `map` or implement it using `bind`?

theories/Core.v
Lysxia: Hm, not sure, I'll try removing the `Monad.` qualfiication.

gmalecha: I was actually thinking of the `_ := fun x =>` rather than `fun _ x =>` or `_ x :=`

theories/Core.v
gmalecha: I was thinking drop it entirely.

Lysxia: Oh right! I didn't see this is in a comment.

Issue #33, opened 2018-11-15
Paco Experiment


gmalecha:
We need to do a paco experiment to determine if it is sufficiently powerful to support the reasoning that we want to do.

Concretely, we decided that proving the unfolding lemma for `mfix` would be a good test. We can probably simplify this a little bit to just proving `interp f (bind c k) = bind (interp f c) (fun x => interp f (k x))`.

gmalecha:
@Lysxia Were you looking into this? Have you gotten a chance to experiment?

gilhur:
Hi all, 

I looked into this project.

The idea of defining 'mfix' is really cool!
I believe that paco should be able to prove everything that is true, but it may require careful thinking and more engineering.
I am happy to help you for this.

One thing I found in the project is that I think the following theorem in `Fix.v` is not true.
```
Theorem mfix_unfold : forall x, mfix x = body E (fun t => id) mfix x.
```
I think one can prove `mfix_unfold -> False` using the classical axiom.

My viewpoint is that `mfix` is a nice syntactic sugar for writing down a definition using `_mfix` and all reasoning should be done in terms of `_mfix`.

Also, I can try to prove the `interp_bind` lemma in MorphismsFacts.v, if you don't mind.
Also, I'd like to talk to you more about this project at POPL!

Thanks,
Gil

gmalecha:
Thanks, Gil! Even a nudge in using Paco would be very appreciated. I'm interested in your comment about the up-to definition, but, regrettably, I haven't had the cycles to look at it lately.

gilhur:
Hi Gregory,

I just finished proving a stronger version of `interp_bind`.
I used the axiom `bisim_eq` to make the proof simpler.
Also, I turned off `Set Primitive Projections` for the definition of `itree` because I don't know how to handle it properly.

Here is my code.
```
From Coq Require Import
     Program
     Classes.RelationClasses
     Classes.Morphisms
     Setoids.Setoid
     Relations.Relations
     Logic.JMeq Logic.EqdepFacts.

Inductive bisimF {E R} (bisim: relation (itree E R)) : relation (itree E R) :=
| Bisim_Ret: forall r, bisimF bisim (Ret r) (Ret r)
| Bisim_Tau: forall t1 t2, bisimF bisim (Tau t1) (Tau t2)
| Bisim_Vis: forall u (e1 e2: E u) k1 k2
                (EvtEQ: e1 = e2)
[...]

PR #40, opened 2018-12-27
paco-based proofs (see #33)

gmalecha:
This is @gilhur's proofs ported to work with the record encoding of positive inductive types.

- there seems to be an issue with `reflexivity` which I can't seem to diagnose.
- `weak_respectful2` doesn't seem to be defined anywhere.


Lysxia:
It looks like Gil's recent PRs subsume this.

PR #41, opened 2018-12-29
Infrastructures and proofs using paco

gilhur:
- Remove "Promitive Projections" for better automation.
- Add infrastructures for coinductive proofs.
- Prove key lemmas.

### DIFF
--- a/theories/Core.v
+++ b/theories/Core.v
@@ -4,7 +4,14 @@ Require Import ExtLib.Structures.Monad.
 
 Set Implicit Arguments.
 Set Contextual Implicit.
-Set Primitive Projections.
+(* Set Primitive Projections. *)
+
+(* Move the following tactic somewhere appropriate *)
+Lemma hexploit_mp: forall P Q: Type, P -> (P -> Q) -> Q.
+Proof. intuition. Defined.
+Ltac hexploit x := eapply hexploit_mp; [eapply x|].
+
+Ltac inv H := inversion H; clear H; subst.
 
 Section itree.
 
@@ -42,12 +49,18 @@ Section itree.
 
    ]]
 
-  *)
+   *)
+
+  Lemma itree_unfold : forall s:itree, s = go (s.(observe)).
+  Proof.
+    intros. destruct s; auto.
+  Qed.
   
 End itree.
 
 Arguments itreeF _ _ : clear implicits.
 Arguments itree _ _ : clear implicits.
+Arguments itree_unfold {E R} s.
 
 (** We introduce notation for the [Tau], [Ret], and [Vis] constructors. Using
     notation rather than definitions works better for extraction.  (The [spin]
--- a/theories/Core.v
+++ b/theories/Core.v
@@ -4,7 +4,7 @@ Require Import ExtLib.Structures.Monad.
 
 Set Implicit Arguments.
 Set Contextual Implicit.
-(* Set Primitive Projections. *)
+Set Primitive Projections.
 
 (* Move the following tactic somewhere appropriate *)
 Lemma hexploit_mp: forall P Q: Type, P -> (P -> Q) -> Q.
@@ -32,6 +32,8 @@ Section itree.
   CoInductive itree : Type := go
   { observe : itreeF itree }.
 
+  Axiom itree_eta : forall s, s = go (s.(observe)).
+
   (** Notes about using [itree]:
 
      - You should simplify using [cbn] rather than [simpl] when working
@@ -51,16 +53,11 @@ Section itree.
 
    *)
 
-  Lemma itree_unfold : forall s:itree, s = go (s.(observe)).
-  Proof.
-    intros. destruct s; auto.
-  Qed.
-  
 End itree.
 
 Arguments itreeF _ _ : clear implicits.
 Arguments itree _ _ : clear implicits.
-Arguments itree_unfold {E R} s.
+Arguments itree_eta {E R} s.
 
 (** We introduce notation for the [Tau], [Ret], and [Vis] constructors. Using
     notation rather than definitions works better for extraction.  (The [spin]
@@ -232,3 +229,9 @@ Global Instance Monad_itree {E} : Monad (itree E) :=
 {| ret := fun _ x => Ret x
 ;  bind := @ITree.bind E
 |}.
+
+Lemma bind'_to_bind {E R U} : forall (t: itree E U) (k: U -> itree E R),
+    ITree.bind' k t = ITree.bind t k.
+Proof. reflexivity. Qed.
+
+Ltac fold_bind := rewrite !bind'_to_bind in *.
--- a/theories/Eq/Eq.v
+++ b/theories/Eq/Eq.v
@@ -210,7 +210,8 @@ Instance eq_itree_vis {E R u} (e: E u) :
   Proper (pointwise_relation _ eq_itree ==>
           @eq_itree E R) (fun k => Vis e k).
 Proof.
-  repeat intro. econstructor; [| econstructor; apply H]; eauto.
+  repeat intro. pfold. econstructor.
+  intros. left. apply H.
 Qed.
 
 Lemma bind_ret {E R} :
--- a/theories/Eq/Eq.v
+++ b/theories/Eq/Eq.v
@@ -52,19 +52,16 @@ Section eq_itree.
     subst. dependent destruction H6. dependent destruction H7. eauto.
   Qed.
 
-  (* Definition eq_itreeF (sim: relation (itree E R)) : relation (itree E R) := *)
-  (*   fun t1 t2 => eq_itreeF' sim (observe t1) (observe t2). *)
-  (* Hint Unfold eq_itreeF. *)
-
-  Notation eq_itreeF :=
-    ((fun sim t1 t2 => eq_itreeF' sim (observe t1) (observe t2)) : relation (itree _ _) -> relation (itree _ _)).
+  Definition eq_itreeF (sim: relation (itree E R)) : relation (itree E R) :=
+    fun t1 t2 => eq_itreeF' sim (observe t1) (observe t2).
+  Hint Unfold eq_itreeF.
 
   Lemma eq_itreeF'_mono : forall x0 x1 r r'
     (IN: @eq_itreeF' r x0 x1) (LE: forall x2 x3, (r x2 x3 : Prop) -> r' x2 x3 : Prop), eq_itreeF' r' x0 x1.
   Proof. pmonauto. Qed.
 
   Lemma eq_itreeF_mono : monotone2 eq_itreeF.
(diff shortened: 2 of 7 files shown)


theories/Core.v
```diff
@@ -42,12 +49,18 @@ Section itree.
 
    ]]
 
-  *)
+   *)
+
+  Lemma itree_unfold : forall s:itree, s = go (s.(observe)).
+  Proof.
+    intros. destruct s; auto.
```
gmalecha: I'm not sure if this is kosher with the new Coq restrictions. https://github.com/coq/coq/pull/7536
It seems like the desire is to have a primitive projection.

theories/Eq/Eq.v
```diff
@@ -5,193 +5,230 @@
 (* TODO: paco-fy this *)
 
 From Coq Require Import
+     Program
```
gmalecha: What is Program used for? I'm always a little bit skeptical of the library because it silently introduces axioms sometimes.

theories/Core.v
gilhur: OK. I can write a version with "Primitive Projections", which is definitely possible though there might be some inconvenience.

theories/Eq/Eq.v
gilhur: "Program" is a standard library.
I just used it to use the tactic "dependent destruction" in a few places.
Not importing "Program" is an easy job, which I will do it.

theories/Core.v
gilhur: I tried it more seriously and it seems that it needs some thinking.
Currently I don't have time to work on this further, so I can do it later this month.
Anyway, I am positive about building a nice infrastructure, so that the current proof structure works almost the same way with "primitive projections".
The main idea is to make every constructor as a morphism for the bismilarity and use the setoid infrastructure to perform `itree_unfold` because `s` and `go (s.(observe))` are bisimilar.

On the other hand, in general, I prefer accepting the bisimilarity axiom and makes everything simpler.
It is exactly the same as the functional extensionality because if you view functions as coinductive data, the bisimilarity coincides with the functional extensionality.
In my mental model, both functional extensionality and bisimilarity axiom work and I can live happily with them.

I wonder what you guys think about this point.

theories/Eq/Eq.v
gilhur: Program is the main library and provides the "Program" feature of Coq.
Currently, I don't use the main feature and only uses some notations and simple ltacs defined there.
Though it is possible to copy and paste some of the notations and ltacs and not import the library, I am not sure whether this is a good idea.
Also as far as I understand, "Program" uses some standard axioms about equality (the axiom K and possibly functional extensionality).

Similarly as before, I am happy with any axioms that are known to be sound and make proof automation better.

theories/Core.v
gmalecha: Personally, I don't really like axioms since they break computation. That said, if people think that it makes things much easier to use, it would be fine. I hadn't thought about the axiom as being the same as functional extensionality, thanks for the insight.

You've already done quite a bit here. I'm sure that someone (maybe myself) can finish porting the proofs based on what you've written here.

On a slightly related note, will Paco need to be updated to meet the bee restrictions on CoInductives in Coq?

theories/Eq/Eq.v
gmalecha: I definitely don't want to propose copying things from the library. Whenever I see it, I just wonder if it is really necessary.

theories/Core.v
gmalecha: For thr morphisms, the thing that we were having problems with was suitably general versions of transitivity and the morohisms to allow rewriting under a pcofix. Perhaps that is addressed here, I will need to look at the PR in more detail.

gilhur: First, in paco, I just added "Set Primitive Projections" and it compiles perfectly.
I will add this in the next update.

Second, setoid does not work well with "eutt", so we need to use the "up to" technique.
But, for the bisimilarity, I think setoid will work well because the coinductive assumption "r" to paco can be always augmented with the greatest respectful closure using the tactic "pupto2_init", which includes the bisimilarity.

Anyway, I can try this later.

Finally, if we simply accept the bisimilarity axiom, we can prove the `itree unfold` lemma even with primitive projections, which will make life easier.

gilhur: I just checked whether my above claim about setoid is correct and unfortunately it turned out that it is wrong.
I think there should be some workaround about this anyway.

However, `itree_unfold` corresponds to the eta expansion for function types, which Coq internalizes as intensional equality.
So, it may make sense to do the same for coinductive types, which would make life much easier.
I will suggest this in [coq/coq#7536](https://github.com/coq/coq/pull/7536).

gilhur:
I just committed a version with primitive projections on but with the itree_unfold (called, itree_eta) axiom.
The update was very mechanical and easy.
The proof scripts remained almost identical.

If the Coq developers accept the eta expansion for coinductive types as intensional equality, all problems are solved.
I suggested this in the discussion post and am waiting to hear from them.

theories/Core.v
gmalecha: I'm interested in understanding exactly what you tried for the bisimilarity and exactly what doesn't work. I have little intuition about this sort of thing and I'd like to understand it better.

As far as itree_unfold, Coq doesn't have definitional eta for *recursive* records, and I imagine that the issue here is similar. With definitional eta for this itree definition, naive conversion would diverge.

gilhur: I don't think the eta rule would introduce any trouble.
For example, the eta rule for functions already works well in Coq.
`f` and `fun x => f x` are treated intensionally identical (e.g. you can prove it by "reflexivity").
I am not sure how it works internally, but my guess is that it reduces only one direction, namely `fun x => f x` to `f`.
Anyway, I think the same idea would work for coinductive types as well because I understand  function types as special coinductive types.

gilhur: What I failed to prove is the following lemma in Eq.v.
```
Lemma eq_itree_clo_bind E R: weak_respectful2 eq_itreeF (@eq_itree_bind_clo E R).
```
The reason why I cannot use the bisimilarity of `itree_unfold` is because the parameter relation "r" given here is an arbitrary relation, not necessarily closed under bisimilarity.
There might be two solutions.
One would be to generalize the respectfulness theory to restrict the relation "r" to be closed under bisimilarity.
The other would be to change the definition of `eq_itree_bind_clo` in an eta expansion form.

But, I think just making Coq to accept the eta equality as is done for function types might be the most natural and easy solution.
Let's wait to hear what Coq developers think about the eta equality.

By the way, I am attending POPL at Lisbon and visiting Boston for 11-15 Feb.
If you are there in either of them, it might be good to have a physical meeting.
Please let me know.
Thanks!

gilhur: Sounds good. Let's talk at POPL further.

Regarding eta equality, let me elaborate more on my thoughts though it may be wrong because I don't know well about how Coq works internally.

What I don't understand from your description is that how the eta expansion can essentially affect reduction behavior.
I looked at some discussions about subject reduction and corecursion, where the problematic expansion is something like `infty = C infty`.
This reduction really introduces one more construct, which can indeed change reduction behavior.

However, eta expansions are not like that.
The reduction `t = go (observe t)` consumes and introduces one construct, so I don't think eta expansions can essentially change any reduction behavior, like eta expansions for functions do not affect reduction behavior (modular finite number of tau steps).

Anyway, Coq already accepts eta equality for functions as definitional equality, so we can better understand this issue by understanding how Coq does it for functions.
Also, at POPL, we might be able to discuss this with Coq developers.

gmalecha: With regards to eta. `observe` is the syntax that forces computation for `cofix`.

```coq
Definition inf := cofix inf := {| Conat := Some inf |}.
```

If you eta-expand `inf`, then you end up with a reduction for the `cofix`. That is, `inf == {| Conat := Some inf |}` holding definitionally (intensionally) which results in divergent computation (at least naively) since `inf` occurs on both sides. Maybe orienting this rule right to left would address the issue, but I'm not sure if it is complete.

This problem doesn't arise in non-recursive records because no subterm can have the same type as the record, so structural recursion  on the type is sufficient for justifying confluence (of this part of things). My understanding is that this is also the reason that eta with functions is ok, i.e. arrows are not recursive.

Lysxia: Hi Gil! These are very interesting discussions, and thanks for the help in finishing those lemmas.

If we introduce `t = go (observe t)`, then let `t = (cofix t := go (TauF t))`, we now have:

```
t = go (observe t)
t = go (observe (cofix t := go (TauF t)))
t = go (observe (go (TauF (cofix ...))))
t = go (TauF (cofix ...))
t = go (TauF t)
```

We seem to end up in a situation similar to the `infty = C infty` problem.

I think the key difference between eta for functions and for coinductive types has to do with the fact that, with functions, the new redex (if there is one) reduces to the original term, so adding eta doesn't significantly change the reduction relation, whereas for coinductive types eta-expansion can introduce entirely new evaluation paths by allowing some previously stuck `cofix` to reduce.

However that's only a problem with making eta definitional. The `itree_unfold` lemma/axiom, or the bisimilarity axiom (which are only propositional) should still be sound as per this Q&A https://github.com/coq/coq/pull/8670/files#r233971706

I think it's worth keeping the bisimilarity axiom for now, even if I'm generally more conservative about introducing axioms.

In any case, reasoning about "up-to-tau" equivalence will put us in front of the same problems and harder ones, and no such axiom will help us there (I believe). Once we get past that, removing the bisimilarity axiom will hopefully be a piece of cake.

gilhur: Oh, I see!
Thanks a lot for the example.
Now I can clearly see that eta-equality can cause a trouble.

OK. Then let's keep the eta-equality axiom for now, which is my latest commit in this PR.
Note that removing the bisimilarity axiom does not introduce much overhead thanks to the up-to technique.

Later we can think about whether there is a way to remove eta-equality without much overhead.

Anyway, please have a look at my proof and let me know if you have any question/comment.

gilhur: > First, in paco, I just added "Set Primitive Projections" and it compiles perfectly.
> I will add this in the next update.

Regarding this, I made a stupid mistake :)
I just added "Set Primitive Projections" without turning the definition into a record form  :)

I just looked into it and since records do not currently support mutual corecursion, I cannot turn the definitions into record forms.

However, this is not a big deal.
Every coinductively defined relations can be defined without using coinduction but instead using impredicativity of Prop.
The idea is to define the greatest fixpoint as the union of all post-fixpoints.
In my original paper, we already implemented Paco in this way as well.

I can easily turn the current Paco not to use any coinductive construction, which I will do later.

gmalecha:
Thanks for the discussion today @gilhur & @Zdancewic . I know that I at least need to walk through the proofs by hand to really understand what is going on, but I'm wondering if we should merge this and try to simplify things more with additional pull requests. What do you guys think?

PR #44, opened 2019-02-02
Remove the eta equality axiom

gilhur:
- Remove the axiom for eta equality.
- Simplify "upto transitivity" into "rewrite".


Lysxia:
I think `unalltaus` and `finite_taus` should still be relations on `itree` rather than `itreeF`. Ideally `itreeF` and `observe` should be internal details that users of the library shouldn't have to reason about.

Lysxia:
I resolved the merge conflict. Let's merge this, and I'll polish things a bit.

Lysxia:
OK I now see why it is also convenient to have versions of `unalltaus` and `finite_taus` on `itreeF` rather than `itree`...

PR #46, opened 2019-02-04
added evaluation operations for locals and memory

Zdancewic:
As a way to take a look at Gregory's Asm implementation, I added semantic handlers for both the local definitions and the memory operations.  We will probably want something like this eventually.


Lysxia:
Seems to be the paco problem that was just resolved in #48

Lysxia:
I can see a way to refactor this part, after reorganizing all the different `interp` functions a bit (so this PR could be merged first):

https://github.com/DeepSpec/InteractionTrees/blob/23a1261f3125da8db754c225ac31fab7ee75ea0a/examples/Asm.v#L149-L158

But the bits above it look like a map library, that could go somewhere in the stdlib or ext-lib, but that seems a bit excessive because these 10 lines are quite readable as they are.

https://github.com/DeepSpec/InteractionTrees/blob/23a1261f3125da8db754c225ac31fab7ee75ea0a/examples/Asm.v#L139-L147

gmalecha:
We already have the ExtLib dependency. I would suggest that we just use the finite maps from that. Though I might be biased.

Zdancewic:
@gmalecha I had the same idea yesterday -- see the new file /theories/EnvEffect.v  (though maybe we should call it "MapEffect"?)

I just pushed changes to this branch that wraps the ExtLib Maps library.  I also needed to use 'interp1' from the new changes to Morphisms.v.   Take a look and let me know what you think.

Zdancewic:
Annoyingly, it looks like the pull request now also includes a bunch of commits that are already present on the master branch.  I thought that I tried to avoid this by following the instructions about how to get git to do that....

If anyone more expert than me at git has a suggestion about how to make this into a "cleaner" pull request, we could do that.  One idea: nuke this branch and make the changes in a new branch (which should be easy since it adds one file and just tweaks Asm.v).

Lysxia:
I can take care of rebasing the branch properly once this is stable to merge.

Lysxia:
Actually, give me a second to rebase this now.

Zdancewic:
OK -- it looks like we can merge this now.

PR #52, opened 2019-02-06
Prove the equivalence between interp and interp1.

gilhur:


Lysxia:
Thanks!

`eutt'` seems to be quite easier to prove things with (whereas `eutt` is easier to consume).

gilhur:
Yes, exactly.
Since the generating functions `eutt_` and `eutt'` are equivalent, conversion between them would be easier.

PR #54, opened 2019-02-08
Threads example

Zdancewic:
An example using extraction and an OCaml "driver" for an itrees program that does round-robin scheduling of three threads.


gmalecha:
This is cool. If I understand correctly, the limitation here is that you can't spawn threads that spawn threads. Is that right?

Zdancewic:
@gmalecha Well, you can spawn threads that spawn threads, but those "internal" threads have to be handled by a different scheduler.  I don't know if it's possible to set things up so that the spawned threads can be handled by the same scheduler.

Lysxia:
(Rebased to integrate the recent paco updates.)

> I don't know if it's possible to set things up so that the spawned threads can be handled by the same scheduler.

The last commit I just added lifts that limitation.

Zdancewic:
Nice!  Shall we merge this to master too?

Issue #21, opened 2018-10-21
Trace Definition


gmalecha:
I've got a few questions about the `is_trace` definition, I'm opening an issue for discussion purposes.

**Inductive vs CoInductive**
Why is `is_trace` `Inductive` rather than `CoInductive`? It seems like coinductive interaction trees should have coinductive traces. We can still use the definitions in `UpTo.v` to approximate the coinductive interaction trees with inductive approximations, and then we can use the current definition of `is_trace` to mostly talk about the the inductive interaction trees.

**Actions that can not Return**
I brought this up the in call on Friday. The current definition of trace doesn't work for interactions that have uninhabited results. For example, the simple effect (in the old style):
```coq
Variant Error : Type -> Type := RuntimeError (_ : string) : Error Empt_set.
```
In the current definition it is impossible to have a trace that includes `RuntimeError`, even if the interaction tree actually calls it, e.g. `Vis (RuntimeError "bad") (fun pf => match pf with end)` (the set of traces that this interaction tree contains only the empty (non-terminating) trace, i.e. `[], None`.

An alternative representation that addresses this problem would be something like:
```coq
CoInductive trace : Type :=
| Tau (_ : trace)
| Send (a : action) (_ : traceK (reaction a))
with traceK (t : Type) : Type :=
| TauK (_ : traceK t)
| Recv (_ : t) (_ : trace).
```
Here, I separate the call and reply actions so that I can represent calls that never return. You could replace both `Tau` and `TauK` with some notion of approximation, though (noting my first comment), I think it makes a bit more sense to do that separately.

A (potential) benefit of the ability to represent this sort of thing is that you should be able to prove a theorem that says that any approximated trace, i.e. trace ending with `None`, can be extended into a longer trace that is consistent with the original `itree`. To have this property, you would either need to add `Tau`s into the trace (like I did above), or add a `Diverge` constructor to both `trace` and `traceK` to represent divergent computations (which are *not* approximated).

Lysxia:
`is_trace`  is `Inductive` to ensure that `spin`  has no nonempty traces. An alternative would be to add `Tau` in traces as you suggest, but then we would have to reason about traces "up to tau". The philosophical motivation for my way of defining `is_trace` is that a trace is what we observe of the program and that `Tau` is not observable.

You are right about the issue with `RuntimeFailure`, and separating the event `{ a : action &  reaction a }` into a `Send` and a `Recv` seems like the right way to make those traces more accurate.

[...]

PR #56, opened 2019-02-08
Update traces to deal with events that do not return

Grain:


theories/Trace.v
```diff
@@ -1,35 +1,237 @@
...
+     Eq.UpToTaus.
+
+From Paco Require Import
+     paco.
 
 Inductive event (E : Type -> Type) : Type :=
-| Event : forall X, E X -> X -> event E.
+| Event : forall X, E X -> X -> event E
+(* An effect without any response from the context (e.g. if X is uninhabited) *)
+| EventOut : forall X, E X -> event E
```
Lysxia: I think we should separate this event from the rest of the trace, to prevent ill-formed traces from being represented. We could have `trace E = list (event E) * option (sigT _ E)` or a custom type like @gmalecha suggested in #21 (this seems a bit more convenient).

Grain: Maybe something like this?
```
Inductive trace' {E : Type -> Type} {R : Type} : Type :=
| TEnd : trace'
| TRet : R -> trace'
| TEventEnd : forall {X}, E X -> trace'
| TEventResponse : forall {X}, E X -> X -> trace' -> trace'.
```
I think @Zdancewic preferred the list definition since working with traces would be easier.

Lysxia:
This looks good to me.

@gmalecha @Zdancewic any more comments?

gmalecha:
This looks pretty reasonable to me. A few questions that are probably worth thinking about before we merge:

1. Is it always the best thing to do to drop `Tau`s from traces? This is probably the case.
2. The separation between `Event` and `EventEnd` seems slightly strange. Originally, I had thought that there would be an `EventOut` and `EventIn` where `EventOut` sends the message to the environment and `EventIn` receives the result. This allows you to, for example, give a modular trace prefix to a simple program that calls the environment. With the definition that you are proposing here, the trace is different if the environment terminates or doesn't terminate.
3. Looking to the future. I wonder a little bit about what the `Event` choice means for concurrent systems. This is by no means a solved question, but in the current definition, you can't really have multiple requests pending on the external environment. This may be desireable, or it may be undesireable, I'm not sure right now.

Closes #21

Zdancewic:
This definition does look reasonable for what we're doing so far, but I agree with @gmalecha that it probably won't scale to concurrency in the way that we might eventually want.  Separating EventIn and EventOut would make things more flexible for future use -- in particular, dealing with asynchronous interactions with the environment.

It is probably worth doing the experiment to see how the EventIn and EventOut traces work out.  

There is the question of what the types of these events should be.  For example, I can imagine that we might want them to be completely symmetric, perhaps something like:

```
Variant event (I O : Type -> Type) :=
| EventIn : forall X, I X -> event I O
| EventOut : forall X, O X -> event I O.

Definition trace I O R := list (event I O) * (option R)
```
For our current purposes, we could define the "type of responses to events E" as something like:

```
Variant responseE {E : Type -> Type} : Type -> Type :=
| respond : forall X, E X -> X -> responseE X.
```

The type of traces we would work with for `itree E R` would be  `trace E (responseE E) R`.

I guess that an alternative to the above would be:
```
Variant event E :=
| EventOut : forall X, E X -> event
| EventIn : forall X, X -> event.
```

For the purposes of our current work, it may be best to just stick with @Grain 's `trace'` for now...

Thoughts?

gmalecha:
What are we planning on using this trace definition for? If what we want to do fits within its expressivity, then we don't need to dig deeper in the immediate future.

Lysxia:
Ok then let's merge this.

- This shows formally an equivalence between itrees and sets of/predicates on traces.

- Predicates on traces can be easier to define (by induction on traces), the trade-off being that you might need to ensure on the side that these predicates are prefix-closed.

- Generalization to support concurrency is future work.

PR #57, opened 2019-02-09
Remove the code from paco-2.0.2

gilhur:


Zdancewic:
Thanks, Gil.  This looks good to me! 

Unless anyone objects, I propose we merge this as soon as possible.

PR #62, opened 2019-02-17
Fixed some typos in the itree predicates example.

Chobbes:
Fixed up a few typos in the predicates example. I think this is quite a good example for this.

An additional nitpick:

```
Theorem state_independent': forall {S R} (t:itree (stateE S) R) 
                            (H: NoGets t),
                             forall s s', ('(s,x) <- interpret_state t s ;; ret x) ≅ ('(s,x) <- interpret_state t s' ;; ret x).
```

I think it should maybe use something like `(newSt, x)` and `(newSt', x)` or something to make it clear that the `s` left of `<-` is different... But it probably doesn't matter much either way!


Lysxia:
Nice!

PR #64, opened 2019-02-18
Cleanup

Zdancewic:


Zdancewic:
Just some slight notation and commenting changes.

Zdancewic:
I plan to do a bit more polishing / cleanup in Eq.v and UpToTau.v next.

YaZko:
Just in case to avoid duplicating some work, be wary that the compiler branch has led to a lot of changes in Eq and UptoTau: both eq_itree and eutt have been parameterized by a relation over the return type, rather than impose equality.

Lysxia:
Some of the renaming here (esp. regarding the use of `_` or `F` suffixes) also overlaps with what took place on the imp2asm branch.

Zdancewic:
Ah -- good to know.  I'll hold off doing more cleanup until we merge that branch, then.

gmalecha:
What is the right thing to do with this PR? Merge it into the Imp2Asm branch?

Lysxia:
I believe the changes have been merged into imp2asm.

PR #66, opened 2019-02-20
Update definition of trace to inductive definition

Grain:
Updated the definition to what I'd like to include in the paper. 

@Lysxia I updated the monad definitions at the bottom which could use a quick look over.

### DIFF
--- a/theories/Trace.v
+++ b/theories/Trace.v
@@ -33,7 +33,8 @@ Fixpoint trace_ret {E R} (tr : @trace E R) : option R :=
   | _ => None
   end.
 
-Inductive is_traceF {E R} : itreeF E R (itree E R) -> @trace E R -> Prop :=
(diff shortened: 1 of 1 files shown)


theories/Trace.v
```diff
@@ -253,10 +235,14 @@ Proof.
   - apply trace_eq_eutt.
 Qed.
 
+Inductive event {E : Type -> Type} : Type :=
+| Event : forall {X}, E X -> X -> event
+.
```
Lysxia: `event`'s argument should be explicit.
```
Arguments event : clear implicits.
```

Grain:
Yes, but I don't think it's too much of a priority now. I didn't mention ```traces``` at all in the paper.

Lysxia:
Ok, sounds good!

Issue #72, opened 2019-03-02
Don't depend on axioms


gmalecha:

gmalecha:
Note that paco still depends on `JMeq_eq`...

gmalecha:
There are definitely lemmas in the latest code that require it, but it might be possible to avoid those lemmas for the actual proofs.

Lysxia:
I had some success in that direction at some point by using heterogeneous equalities when comparing `VisF` constructors. Essentially the situation is that `inversion` sometimes gives you heterogeneous equalities, but you still need to prove homogeneous equalities. What I'm proposing weakens those goals while keeping the same hypotheses.

For example this `eq_notauF` constructor

```
| Eutt_vis : forall u (e : E u) k1 k2,
    (forall x, euttE (k1 x) (k2 x)) ->
    eq_notauF euttE (VisF e k1) (VisF e k2).
```

should be

```
| Eutt_vis : forall u1 (e1 : E u1) u2 (e2 : E u2) k1 k2,
    eq_dep u1 e1 u2 e2 ->
    (forall x1 x2, JMeq x1 x2 -> euttE (k1 x1) (k2 x2)) ->
    eq_notauF euttE (VisF e1 k1) (VisF e2 k2).
```

gmalecha:
Whenever I see JMeq, I figure that it is probably going to force me to use an axiom at some point. Are you able to do proofs with this without an axiom?

Lysxia:
I think we can at least make it so that this library doesn't introduce axioms on its own, so they should only come from paco, which will be fixed in its own time.

> Are you able to do proofs with this without an axiom?

I could prove transitivity of `eutt`. I don't remember whether I did anything substantial with `bind`. `JMeq` and `eq_dep` enjoy their flavors of transitivity, etc. which made things okay.

@gmalecha Are you working on this?

Lysxia:
We can actually take this issue off the first milestone.

gmalecha:
Yes, this can be removed from the first milestone. I put it down during the big refactoring and I  haven't gotten back to it.

Lysxia:
This is currently a low priority issue (proof irrelevance is not too bad). Fixing paco first may be a good starting point for this work.

PR #75, opened 2019-03-03
Try to remove reliance of `proof_irrelevance`.

gmalecha:
There are definitely lemmas that we can not remove it from. What remains to be seen is whether we truly need those lemmas.


gmalecha:
I think something like this is worth merging, but possibly not in it's current form. Should we postpone until the re-org is done?

Lysxia:
Sure!

Lysxia:
Closing as stale. We can keep discussing this in #72

Issue #73, opened 2019-03-02
Determine if we can make interp equivalent to translate when using `Vis e Ret`


gmalecha:
This would require that we only introduce a `Tau` constructor for `Vis` if the effect interpretation is `Ret`.

gmalecha:
I gave this a try and the definition isn't that much more complicated. I need to understand a bit more about paco proofs to update the lemmas though.

```coq
Definition interp {E F : Type -> Type} (h : E ~> itree F) :
  itree E ~> itree F := fun R =>
  cofix interp_ t :=
    handleF interp_
            (fun _ e k =>
               match (h _ e).(observe) with
               | RetF r => Tau (interp_ (k r)) (* only insert Tau if you are interpreting immediately to a Ret *)
               | TauF z => Tau (ITree.bind z (fun x => interp_ (k x)))
               | VisF e k' => Vis e (fun x => ITree.bind (k' x) (fun x => interp_ (k x))) (* don't inject the extra Tau *)
               end)
            (observe t).
```

This this means that `eq_itree (translate f it) (interp (fun _ e => Vis (f _ e) (fun x => Ret x)) it)` which would allow us to change this equation into a definition. As long as we reason using equational rewriting, there shouldn't be any impact on users.

gmalecha:
@Lysxia Any initial thoughts on this? I think the proofs for `interp` would be a little bit more complex, but we would get all the `translate` proofs "for free".

Lysxia:
This looks reasonable, for sure.

Lysxia:
I gave that a quick try but couldn't figure out how to fix the proofs.

You can repackage this `match` into a `bind_delay`, which is equivalent to `bind`, but inserts a `Tau` if `Ret` is the first constructor.

Lysxia:
I now think that, as far as this library is concerned,  `Tau`'s are too finicky so the library should try to hide `Tau` (and paco) behind `eutt` as much as possible. This seems to add too much complexity for the ability to use  `eq_itree` or `eq` in non-visible ways. So let's close this, but I'm open to further related suggestions.

PR #79, opened 2019-03-06
Use aloop as a primitive + refactoring

Lysxia:
Several changes orthogonal to each other, so they can be split if required, but that's more work.

1. The main change is to use (an updated) `aloop` to implement all interpreters. Ideally the only `CoFixpoint` left with this are `bind` and `aloop`, though there are still some leftovers at the moment.

    1. As a consequence, outside of this little core, proofs do not rely on the reduction semantics of cofixpoints.

    2. It might be possible to reduce this further to a single cofixpoint.

    3. The benefits of this are certainly debatable. This mostly affects some internals of the library. From the outside, this changes little as long as all Cofixpoint come with associated unfolding lemmas (which just pushes the `cofix` deeper). And even within the library, in principle the change should only affect the proofs of these unfolding lemmas, which everything else should rely on. So I'm not sure whether reducing the number of cofixpoints inside the library is something to strive for.

2. Remove `interp1`. (Partly to reduce work, but there seems to be some agreement that it makes things more complicated than they need to be.)

3. Simplify proofs. In particular, I got rid of a lot of occurences of `itree_eta`. Most of the time you can use an unfolding lemma (this could also remove the need for all lemmas of the style `ret_bind`, `vis_bind`, `tau_bind`, but maybe they're still convenient to keep around in any case).

### DIFF
--- a/examples/Imp2AsmCorrectness.v
+++ b/examples/Imp2AsmCorrectness.v
@@ -342,7 +342,7 @@ Section Correctness.
 
   Definition interp_locals {R: Type} (t: itree E R) (s: alist var value)
     : itree E' (alist var value * R) :=
-    run_env _ (interp1 evalLocals _ t) s.
+    run_env _ (interp (eh_both evalLocals eh_inr) _ t) s.
 
   Instance eutt_interp_locals {R}:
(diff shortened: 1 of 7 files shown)


Lysxia:
There are some cofixpoints left but let's merge this for now.

gmalecha:
I will work on #73, that would eliminate one more `cofix`.

Issue #86, opened 2019-03-06
Use the coqdoc infrastructure


Zdancewic:
But, keep in mind the SF markup stuff for the future.

gmalecha:
I have used https://www.ps.uni-saarland.de/~ttebbi/coqdocjs/ and it produces pretty nice output.

PR #94, opened 2019-03-12
try out coqdocjs

Zdancewic:
This pull request addresses issue #86  

- adds coqdocjs as a git submodule  (so we can use it for post-processing coqdoc output)
- adds an 'html' target to the Makefile

Do we want to add the coqdocjs dependency?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,6 +29,7 @@ Makefile.coq.conf
 _CoqProject
 _CoqPath
 lib/
+html/
 
 tests/extraction/*.ml
 tests/extraction/*.mli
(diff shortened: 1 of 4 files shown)


Lysxia:
I think coqdocjs is cool, but we probably shouldn't track it as a submodule. We can add a few lines in the README to explain how to install it.

Zdancewic:
OK -- I adjusted the pull request to make the coqdocjs stuff optional and updated the README.md accordingly.

Lysxia:
It looks like the `.gitmodules` file can be removed.

Zdancewic:
OK -- I removed the .gitmodules file and did the merge.

PR #96, opened 2019-03-13
split out recursion from interp, raise level of abstraction

Zdancewic:
This pull request does two (related things).  It proposes to split out Recursion and RecursionFacts from Interp and InterpFacts.  I think that this is cleaner because Handler.v depends on Interp but not Recursion.  Also, once we have handlers (and their category) available, we can raise the level of abstraction for Recursion.  In particular, we can have `interpret_recursive_call`.

Step through the rewriting in examples/Factorial.v  (which no longer relies on itree_eta) for an example of how looks -- in particular, we never need to see the internal plumbing of the coproduct.


Lysxia:
Shall we merge this? LGTM.

PR #97, opened 2019-03-14
Add rec-fix and mrec-fix notations

Lysxia:
Notations that resemble the `fix` construct


Zdancewic:
This looks good to me.  I think we should merge it.

PR #99, opened 2019-03-15
Rename many lemmas with more consistent names

Lysxia:


gmalecha:
The naming scheme in ExtLib is roughly "reading from left to right, the symbols that you see separated by _".

Lysxia:
Isn't it tricky when notations like `>>=`/`_ <- _ ;; _` are involved?

Issue #83, opened 2019-03-06
Some proofs are slow


Lysxia:
With the further split of some `eq_itree` properties into `Eq/Shallow.v` (typically as `observing eq`), it sometimes takes up to 10s to use those "shallow" equivalences for rewriting. Some examples of this are marked as TODO.

The `observing` pattern is common enough that it would be a shame to drop this module for performance reasons.

gmalecha:
Can you reference specific tactic lines that take a long time? I might have some thoughts about them.

Lysxia:
You can take a look at `rewrite ret_bind.` in `Morphisms.unfold_interp`, and `rewrite unfold_bind.` in `Morphisms.interp_bind`.

Lysxia:
It looks like these quoted proofs are no problem now. I'll close this and open more specific issues for other slow parts.

PR #109, opened 2019-03-24
Make some rewrites faster

Lysxia:
```
(* slower *)
Instance eq_f r : Proper (sig) (f r).

(* faster *)
Instance eq_f : Proper (eq ==> sig) f. 
```

(It seems `Proper` instance resolution is designed to peel off arguments first)

```
(* slower *)
Instance Proper_eutt : Proper (sig ==> flip impl) eutt.

(* faster *)
Instance Proper_eutt : Proper (sig ==> iff) eutt.
```

(`impl` and `flip impl` are obvious subrelations of `iff`, whereas deriving a `Proper` instance for `impl` from a `flip impl` take more work.)

There are still quite a few slow places. e.g., `unfold_bind` in `Effects.StateFacts`.

### DIFF
--- a/theories/Effects/StateFacts.v
+++ b/theories/Effects/StateFacts.v
@@ -5,9 +5,6 @@ From Coq Require Import
      Morphisms
      RelationClasses.
 
-From ExtLib Require Import
-     Structures.Monoid.
-
 From Paco Require Import paco.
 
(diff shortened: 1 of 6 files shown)


Lysxia:
Related to #83

PR #121, opened 2019-04-01
Update tutorial to talk about mapE instead of envE.

Chobbes:
Noticed that the tutorial had comments which were slightly out of date.


YaZko:
Good catch, thanks!

PR #132, opened 2019-06-02
Remove the tactic `auto_inj_pair2`

gilhur:
I replaced `inv` by `dependent destruction` when the tactic `auto_inj_pair2` is needed.

### DIFF
--- a/theories/Eq/Eq.v
+++ b/theories/Eq/Eq.v
@@ -443,7 +443,7 @@ Qed.
 Lemma eq_itree_inv_vis {E R U} (t : itree E R) (e : E U) (k : U -> _) :
   t ≅ Vis e k -> exists k', observe t = VisF e k' /\ forall u, k' u ≅ k u.
 Proof.
-  intros; punfold H; inv H; auto_inj_pair2; subst; try inv CHECK.
+  intros; punfold H; red in H; dependent destruction H.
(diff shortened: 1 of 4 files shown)


Lysxia:
Fair enough.

That reminds me that it's possible to avoid axioms (which `dependent destruction` depends on) altogether in the itree library (by using `eq_dep` in the `VisF` rule of what is now `eqit`), but paco already uses `JMeq_eq`. Is there any way to make paco axiom-free as well?

gmalecha:
I recall the idea of axiom-free paco being brought up a long time ago.

gilhur:
Oh I forgot about the axiom issue.
Here is a summary.
- "JMeq_eq" is equivalent to UIP (Uniqueness of identity proof), which may be problematic in Homotopy Type Theory.
- Paco uses "JMeq_eq" only for implementing the tactics.
  I guess it might be possible to implement them without using "JMeq_eq".
- I already used "dependent destruction" in many places and that's why I removed "auto_inj_pair2".

But, since we may want to try to remove "JMeq_eq" later, let's keep the tactic "auto_inj_pair2".

I will close this pull request.

Lysxia:
@gilhur `auto_inj_pair2` also uses an axiom, so doesn't this PR still make sense to make the proofs more uniform?

PR #134, opened 2019-06-10
Simplify euttG

gilhur:
I simplified the definition of `euttG` as follows:
- from `gpaco2 ... (transU rH \2/ transD rL) (transU rH \2/ transD rL \2/ transD gL)`
- to `gpaco2 ... (transU rH \2/ rL) gL`


YaZko:
Ah those look great, thanks Gil!

PR #145, opened 2019-11-10
[WIP] Fixes for compatibility with Coq v8.10 and master

tchajed:
I've tested this on Coq v8.9.1, v8.10.1, and the current Coq master.

An incomplete list of changes needed:

- Coq now generates slightly different files, which requires new gitignore rules.
- Instances can no longer be partially provided with `:=` (so that it's syntactically obvious whether or not we're in proof mode). The relevant change to Coq for InteractionTrees is https://github.com/coq/coq/pull/9270.
- The projection syntax now only supports record projections (see https://github.com/coq/coq/pull/8829). This is a bit sad since it would be nice for the user to think of `observe` as a projection; hopefully Coq supports that in some way.

There was one mysterious breakage in `Eq/UpToTaus.v`. The rewrite previously there now goes into an infinite loop while resolving typeclasses for setoid rewriting. I managed to fix it by essentially manually giving the right instance, but it would be nice to fix the underlying issue.

### DIFF
--- a/theories/Core/ITreeDefinition.v
+++ b/theories/Core/ITreeDefinition.v
@@ -241,13 +241,13 @@ Module ITreeNotations.
 Notation "t1 >>= k2" := (ITree.bind t1 k2)
   (at level 50, left associativity) : itree_scope.
 Notation "x <- t1 ;; t2" := (ITree.bind t1 (fun x => t2))
-  (at level 60, t1 at next level, right associativity) : itree_scope.
+  (at level 61, t1 at next level, right associativity) : itree_scope.
 Notation "t1 ;; t2" := (ITree.bind t1 (fun _ => t2))
-  (at level 60, right associativity) : itree_scope.
+  (at level 61, right associativity) : itree_scope.
 Notation "' p <- t1 ;; t2" :=
   (ITree.bind t1 (fun x_ => match x_ with p => t2 end))
-  (at level 60, t1 at next level, p pattern, right associativity) : itree_scope.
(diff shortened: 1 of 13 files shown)


tchajed:
Oops, this seemed to work but I ran into another rewrite infinite loop in `tutorial/AsmCombinators.v`.

theories/Core/ITreeDefinition.v
```diff
@@ -241,13 +241,13 @@ Module ITreeNotations.
...
+  (at level 61, t1 at next level, right associativity) : itree_scope.
 Notation "t1 ;; t2" := (ITree.bind t1 (fun _ => t2))
-  (at level 60, right associativity) : itree_scope.
+  (at level 61, right associativity) : itree_scope.
 Notation "' p <- t1 ;; t2" :=
   (ITree.bind t1 (fun x_ => match x_ with p => t2 end))
-  (at level 60, t1 at next level, p pattern, right associativity) : itree_scope.
-Infix ">=>" := ITree.cat (at level 60, right associativity) : itree_scope.
+  (at level 61, t1 at next level, p pattern, right associativity) : itree_scope.
+Infix ">=>" := ITree.cat (at level 61, right associativity) : itree_scope.
```
Lysxia: The file `coq-itree.opam` should also be updated with the new version of coq-ext-lib because there's a conflict somewhere in the examples or the tutorial.

Lysxia:
I ended up fixing compatibility with 8.10 independently in #153 (from a branch that was dormant prior to this PR). I'll borrow your first commit for the gitignore as part of adding Coq master to CI (WIP on branch coq-master). Thanks in any case!

PR #151, opened 2019-12-27
Dijkstra monad iter pred refactor

lag47:
Merging in refactor of dijkstra monad material


lag47:
Merging in refactor of dijkstra monad material

PR #157, opened 2020-01-17
Fix compatibility with ExtLib

liyishuai:


tutorial/Imp2Asm.v
```diff
@@ -22,6 +22,7 @@ Require Import Imp Asm Fin Utils_tutorial AsmCombinators.
 Require Import Psatz.
 
 From Coq Require Import
+     List
```
Lysxia: Is this necessary?

liyishuai: Yes for future versions of Coq.

theories/Core/ITreeDefinition.v
```diff
@@ -239,7 +239,7 @@ End ITree.
 
 Module ITreeNotations.
```
Lysxia: I'd like to try removing these notations altogether, so we can stop messing with the coq-ext-lib versions.

However the `PMonad` removal means that definitions are going to change, so we might have to stick to 0.11.1 this time anyway...

tutorial/Imp2Asm.v
Lysxia: 8.11? What breaks?

liyishuai: https://circleci.com/gh/coq-community/coq-ext-lib/163
```
# File "./Utils_tutorial.v", line 624, characters 21-33:
# Error: The reference ascii_of_nat was not found in the current environment.
```

theories/Core/ITreeDefinition.v
Lysxia: I was hoping I could do a 2.1.0 with minimal breaking changes but there's already the change in notation from 0.10.3, so I'm going to make the next release 3.0.0 anyway. In that case we might as well take this PR and require upgrading to coq-ext-lib 0.11.1.

Lysxia:
LGTM but CI should pass before merging.

Lysxia:
I keep forgetting about the `PMonad` removal, which is a pretty major change.

I'll first stick to 0.10.3 for the next release, and we'll upgrade to 0.11.1 soon after by merging this PR.

Lysxia:
Let's merge this once tests pass.

PR #159, opened 2020-01-29
[.travis.yml] Test out with OCaml 4.09

artagnon:
Most users will probably have this configuration, so let's put it in Travis.


Lysxia:
That's fair. Thanks!

PR #171, opened 2020-06-03
Fixing the definition of mrec_fix.

### LINKED ISSUE #170: Unexpected definition of `mrec_fix`
I find the current definition of `mrec_fix` quite surprising: https://github.com/DeepSpec/InteractionTrees/blob/master/theories/Interp/Recursion.v#L103
It has two parameters `A` and `B` that are unused.
Furthermore the notation `mrec-fix` defined below is putting a wildcard for the type `T`, which is typically needed for mutually recursive definitions.

Here is a proof script to illustrate my point.

The definition `odd_even_1` is made using `mrec` and works. Note the importance of using the argument `T` (which why I was so surprised to find a wildcard in `mrec-fix`).

The natural definition `odd_even_2` is failing because of the wildcard. The definition `odd_even_3` suggests a fix.

```coq
From ITree Require Import ITree ITreeFacts.

Inductive odd_evenE : Type -> Type :=
  | odd : nat -> odd_evenE bool
  | even : nat -> odd_evenE bool
  .

Definition odd_even_1 : odd_evenE ~> itree void1 :=
  mrec (fun T (d : odd_evenE T) =>
    match d with
    | odd 0 => ret false
    | odd (S m) => trigger (even m)
    | even 0 => ret true
    | even (S m) => trigger (odd m)
    end).

Fail Definition odd_even_2 : odd_evenE ~> itree void1 :=
  mrec-fix odd_even d :=
    match d in odd_evenE T return itree (odd_evenE +' void1) T with
    | odd 0 => ret false
    | odd (S m) => odd_even _ (even m)
    | even 0 => ret true
    | even (S m) => odd_even _ (odd m)
    end.

Local Notation endo T := (T -> T).

Definition mrec_fix2 {D E : Type -> Type}
           (ctx : endo (D ~> itree (D +' E)))
  : D ~> itree E
  := mrec (ctx trigger_inl1).

Notation "'mrec-fix2' f d := g" :=
  (let D := _ in
   mrec_fix2 (D := D) (fun (f : forall T, D T -> _) T (d : D T) => g))
  (at level 200, f ident, d pattern).

Definition odd_even_3 : odd_evenE ~> itree void1 :=
  mrec-fix2 odd_even d :=
    match d in odd_evenE T return itree (odd_evenE +' void1) T with
    | odd 0 => ret false
    | odd (S m) => odd_even _ (even m)
    | even 0 => ret true
    | even (S m) => odd_even _ (odd m)
    end.
```

Mbodin:
Fixes #170.

### DIFF
--- a/theories/Interp/Recursion.v
+++ b/theories/Interp/Recursion.v
@@ -100,12 +100,14 @@ Arguments trigger_inl1 {D E} [T].
 (** Short for endofunctions, used in [mrec_fix] and [rec_fix]. *)
 Local Notation endo T := (T -> T).
 
-Definition mrec_fix {D E : Type -> Type} {A B : Type}
+Definition mrec_fix {D E : Type -> Type}
            (ctx : endo (D ~> itree (D +' E)))
   : D ~> itree E
   := mrec (ctx trigger_inl1).
 
-Notation "'mrec-fix' f d := g" := (mrec_fix (fun f _ d => g))
+Notation "'mrec-fix' f d := g" :=
+	(let D := _ in
+	 mrec_fix (D := D) (fun (f : forall T, D T -> _) T (d : D T) => g))
   (at level 200, f ident, d pattern).
 (* No idea what a good level would be. *)
 


Lysxia:
Thanks a lot for this fix!

I wonder whether it would be possible to have a solution that survives simplification, but I couldn't figure it out; I'll keep track of that in another issue.

PR #210, opened 2021-05-27
Cartesian closure

Zdancewic:
I've been doing a bit of work extending the category theory base to include cartesian closed categories.  This pull request adds the basic components for products and exponentials and instantiates them for `Fun`.  I'm not 100% sure about how the global vs. local typeclass instances for products interacts with the way I've specified the typeclasses for curry and apply.  I also imagine that there are more standard lemmas about CCC's that could be added to `CategoryFacts.v`, but I'm not sure where to look for a relevant list of lemmas, but this is a starting point.

### DIFF
--- a/theories/Basics/CategoryTheory.v
+++ b/theories/Basics/CategoryTheory.v
@@ -306,7 +306,7 @@ Class CartesianClosed : Prop := {
   cartesian_product :> Product _ PROD;
   cartesian_curry_apply :> CurryApply;
   cartesian_proper_curry_ :> forall a b c,
-    @Proper (C (PROD c a) b -> C c (EXP b a)) (eq2 ==> eq2) curry_                          
+    @Proper (C (PROD c a) b -> C c (EXP a b)) (eq2 ==> eq2) curry_                          
 }.
   
 End CartesianClosureLaws.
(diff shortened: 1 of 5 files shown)


theories/Basics/CategoryTheory.v
```diff
@@ -269,6 +279,40 @@ Arguments pair_fst {obj C Eq2_C Cat_C bif Pair_C Fst_C PairFst} [a b c] f g.
...
+Context {obj : Type} (C : Hom obj).
+Context {Eq2_C : Eq2 C} {Id_C : Id_ C} {Cat_C : Cat C}.
+Context (ONE : obj)
+        {Term_C : Terminal C ONE}.
+Context (PROD : binop obj).
+Context {Pair_C : Pair C PROD}
+        {Fst_C : Fst C PROD}
+        {Snd_C : Snd C PROD}.
+Context (EXP : binop obj).
+Context (Apply_C : Apply C PROD EXP)
```
Zdancewic, line 294: Now that I look at this, the `Apply_C` and `Curry_C` contexts should probably be implicit (in analogy with the other operations).

Lysxia:
This looks all good to me!

PR #216, opened 2021-11-04
Trace

lag47:
One of 3 partial commits to bring the secure branch into master. This branch adds the ITraces, a linear, coinductive trace data type built on top of ITrees. ITraces are implemented as ITrees over a specialized event type family. This branch introduces a trace refinement relation, creating a trace model for ITrees that is proven sound and complete over eutt.

(Note that the Coq 8.13 + make build is breaking on this pull request. However, the dijkstra_pull_request and secure branches do not have this problem, so it is probably not worth tracking down)

### DIFF
--- a/theories/ITrace/ITraceFacts.v
+++ b/theories/ITrace/ITraceFacts.v
@@ -1,0 +1,724 @@
+From Coq Require Import
+     Morphisms.
+
+From ITree Require Import
+     Axioms
+     ITree
+     ITreeFacts
+     Eq.Rutt
+     Props.Divergence
+     Props.EuttDiv
+     ITrace.ITraceDefinition
+.
+
+
+From Paco Require Import paco.
+
+Import Monads.
+Import MonadNotation.
+Local Open Scope monad_scope.
+
+Lemma classic_empty : forall (A : Type), ( exists e : A + (A -> void), True ).
+Proof.
+  intros. destruct (classic (exists a : A, True)).
+  - destruct H; eauto.
+  - assert (f : A -> void); eauto. intros.
+    exfalso. apply H; eauto.
+Qed.
+
+Lemma may_converge_trace : forall (E : Type -> Type) (R : Type)
+                                   (b : itrace E R) (r1 r2 : R),
+    may_converge r1 b -> may_converge r2 b -> r1 = r2.
+Proof.
+  intros. induction H; inversion H0; subst.
+  - rewrite H in H1. pinversion H1. subst. auto.
+  - rewrite H in H1. pinversion H1.
+  - destruct e. destruct b. apply IHmay_converge. rewrite H in H0. inversion H0; subst;
+                                                                    contra_void.
+    + pinversion H3.
+    + destruct e; [ | contradiction ]. destruct b.
+      pinversion H3. ddestruction.
+      enough (k tt ≈ k0 tt); try apply REL.
+      rewrite H5. auto.
+    + contradiction.
+ - destruct e. destruct e0. destruct b. destruct b0.
+   apply IHmay_converge. rewrite H in H2.
+   pinversion H2. ddestruction.
+   subst. enough (k tt ≈ k0 tt); try apply REL.
+   rewrite H4. auto; contra_void.
+   + destruct b0.
+   + destruct b.
+Qed.
+
+Lemma finite_nil {E : Type -> Type} : finite (@Nil E).
+Proof.
+  apply conv_ret. unfold Nil. reflexivity.
+Qed.
+
+Lemma finite_list_to_stream : forall (E : Type -> Type) (l : ev_list E),
+    finite (ev_list_to_stream l).
+Proof.
+  red. intros. induction l.
+  - cbn. constructor. reflexivity.
+  - cbn. eapply conv_vis; try reflexivity. simpl. eauto.
+    Unshelve. exact tt.
+Qed.
+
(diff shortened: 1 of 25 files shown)


theories/Basics/Basics.v
```diff
@@ -96,6 +96,7 @@ Polymorphic Class MonadIter (M : Type -> Type) : Type :=
 
     Quite easily in fact, no [Monad] assumption needed.
  *)
+(** I needed to make a change for this to typecheck*)
```
Lysxia, line 99: ?

theories/ITrace/EuttEv.v
```diff
@@ -0,0 +1,231 @@
...
+  | EqRet : forall (r1 : R1) (r2 : R2), RR r1 r2 -> euttEvF vclo sim (RetF r1) (RetF r2)
+  | EqTau : forall (m1 : itree E1 R1) (m2 : itree E2 R2), sim m1 m2 -> euttEvF vclo sim (TauF m1) (TauF m2)
+  | EqTauL : forall (t1 : itree E1 R1) (ot2 : itree' E2 R2),
+      euttEvF vclo sim (observe t1) ot2 -> euttEvF vclo sim (TauF t1) ot2
+  | EqTauR : forall (ot1 : itree' E1 R1) (t2 : itree E2 R2),
+      euttEvF vclo sim ot1 (observe t2) -> euttEvF vclo sim ot1 (TauF t2)
+  | EqVis : forall (A B : Type) (e1 : E1 A) (e2 : E2 B ) (k1 : A -> itree E1 R1) (k2 : B -> itree E2 R2),
+      REv e1 e2 -> 
+      (forall (a : A) (b : B), RAns e1 a e2 b -> vclo sim (k1 a) (k2 b) : Prop) -> euttEvF vclo sim (VisF e1 k1) (VisF e2 k2).
+(*what if RE is empty? this definition seems to have serious problems*)
```
Lysxia, line 56: What is `RE`? Is this comment still up to date?

theories/ITrace/EuttEv.v
```diff
@@ -0,0 +1,231 @@
...
+Qed.
+
+
+
+Lemma euttEv_id_monot : monotone2 (@id (itree E1 R1 -> itree E2 R2 -> Prop) ).
+Proof. auto. Qed.
+ 
+
+
+Definition euttEv : itree E1 R1 -> itree E2 R2 -> Prop := paco2 (euttEv_ id) bot2.
```
Lysxia, line 80: That is a pretty general definition that probably deserves to go under `Eq/`. Would it also sound better to rename it `rutt`/`relutt` (R for relation) or `hutt` (H for heterogeneous)?

theories/ITrace/ITraceFacts.v
```diff
@@ -0,0 +1,840 @@
...
+- apply (Tau (determinize_ E R (observe t) ) ).
+- specialize (classic_empty X) as H. 
+  apply constructive_indefinite_description in H. destruct H as [ [x | f] _].
+  + apply (Vis (evans X e x) (fun _ =>  (determinize_ E R (observe (k x)) ) )) .
+  + apply (Vis (evempty X f e) (fun v : void => match v return itrace E R with end)  ).
+Defined.
+
+Definition determinize {E R} t := determinize_ E R (observe t).
+
+(* may be a better idea to make this an axiom *)
```
Lysxia, line 442: I'm pretty sure this is equivalent to LEM, so the current set up on top of LEM is less surprising.

Lysxia:
This looks good to me! Just a few comments above to address, and I need to update the documentation about axioms.

theories/ITrace/EuttEv.v
lag47: This comment is not up to date and I can just delete it

theories/ITrace/EuttEv.v
lag47: I'm fine with it being renamed and being moved to `Eq/`. I'm not sure I'm crazy about either suggested name, because I don't see that they make it clear that the heterogeneity of this relation has to do specifically with the events. But I'm not sure I have a better suggestion.

Lysxia: I'm partial to "relation" as a reference to Reynold's "types as relations" interpretation of parametricity, which generally maps type constructors to heterogeneous relations, and this is more or less an instance of it.

theories/ITrace/ITraceFacts.v
lag47: I believe the axiom that I use `slem : forall (P : Prop), {P} + {~P}` is strictly stronger than `lem : forall (P : Prop), P \/ ~ P` . This is mentioned in this link [https://github.com/coq/coq/wiki/CoqAndAxioms](https://github.com/coq/coq/wiki/CoqAndAxioms). Basically the strong law of the excluded middle allows you to write functions like 
```coq
if P x then e1 else e2
``` 
for arbitrary, non decidable, P. This would break extraction for any function that uses slem. I only think that is ok because it is only used to create a coinductive value that is only used for to be a witness in a proposition.

Lysxia: The axiom of (countable) choice might be a sufficient standard axiom. But anyway we can leave figuring out this question as future work.

Lysxia:
I also moved `eutt_div`, `must_diverge` and `can_converge` into `Props.EuttDiv` and `Props.Divergence` since they can be defined without reference  to traces.

PR #217, opened 2021-11-04
Dijkstra pull request

lag47:
One of three pull requests intended to bring the secure branch into master.  This branch introduces the main body of code implementing the [Dijkstra Monads Forever](https://dl.acm.org/doi/10.1145/3434307) paper. This includes a formalization of Dijkstra Monads and their laws as a type classes, the StateSpecT state transform over Dijkstra Monads, and TraceSpec a Dijkstra Monad over ITrees with arbitrary uninterpreted events built on top of ITraces.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,7 +30,9 @@ deps.dot
 .coqdeps.d
 Makefile.coq
 Makefile.coq.conf
-_CoqProject
+/_CoqProject
+examples/_CoqProject
+tutorial/_CoqProject
 _CoqPath
 lib/
 html/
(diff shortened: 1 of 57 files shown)


Lysxia:
Here's a general way to avoid axioms in `Type` like  strong excluded middle: wrap it in `inhabited`, e.g., `classicT_inhabited : inhabited (forall T, T + (T -> False))`. When defining (noncomputable) functions that actually need it, take it as a parameter. Then in proofs (that are in `Prop`), we can first `destruct classicT_inhabited as [classicT]`, and use the given parameter in to call those noncomputable functions.

Lysxia:
Another trick I found is that if we make `resp_eutt` a notation for `Proper (eutt ==> iff)`, then we can simply `rewrite` instead of applying these assumptions manually.

PR #220, opened 2021-12-19
rutt: swapping order of constructors and couple of simplifications

YaZko:
Minor simplifications to [rutt]:

- Swapped the order of constructors so that they match the definition of [eqit]
- Removed the [vclo] argument: it is introduced in [eqit] exclusively for the definition of [euttG]. If we don't define something similar here, it is completely useless
- Specialized the up-to [eqit] closure. It is only useful to prove up-to [euttge] so no need to carry all these booleans around.

Note: the up-to [eqit] closure is almost a duplicate of the one used to close [eutt] itself, except that this one is heterogeneous in [E]. We could generalize the latter and reuse it here to avoid some duplication.

Also fixed a bunch of warnings against 8.14, hope it doesn't break compatibility with older versions of Coq.

### DIFF
--- a/theories/Eq/Rutt.v
+++ b/theories/Eq/Rutt.v
@@ -25,7 +25,6 @@ From ITree Require Import
      ITreeFacts
 .
 
-
 From Paco Require Import paco.
 
 Import Monads.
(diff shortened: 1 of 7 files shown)


Lysxia:
There's a missing lower bound on coq in coq-paco.dev https://github.com/coq/opam-coq-archive/pull/2002

and I've checked that it still compiles on 8.10 on my machine.

theories/Eq/Rutt.v
```diff
@@ -177,44 +173,46 @@ Proof.
...
-Hint Resolve (eqitC_rutt_wcompat') : paco.
+#[export] Hint Resolve euttge_trans_clo_wcompat : paco.
 
-Global Instance grutt_cong_eqit {R1 R2 : Type} {E1 E2 : Type -> Type} {REv : forall A B, E1 A -> E2 B -> Prop}
-      {RAns : forall A B, E1 A -> A -> E2 B -> B -> Prop} {RR1 RR2} {RS : R1 -> R2 -> Prop} r rg
-      (LERR1: forall x x' y, (RR1 x x': Prop) -> (RS x' y: Prop) -> RS x y)
+(* The validity of the up-to [euttge] entails we can rewrite under [euttge]
+   and hence also [eq_itree] during coinductive proofs of [rutt]
+*)
+#[global] Instance grutt_cong_eqit {R1 R2 : Type} {E1 E2 : Type -> Type} {REv : forall A B, E1 A -> E2 B -> Prop}
```
Lysxia, line 198: I am grutt.

Lysxia (approved):

Lysxia:
I'll just remove 8.10 and 8.11 from life support.

liyishuai:
Need to update `coq-itree.dev` in opam-coq-archive accordingly.

Lysxia:
Oh, right, I can do that!

PR #232, opened 2022-03-04
ci: Add 8.15

liyishuai:


bcpierce00:
@liyishuai Looks like one of the CI jobs is stuck?

Lysxia:
It looks like the build failed on CircleCI but it forgot to tell Github. Also the failure is just Coq dev breaking itree, which happens about every month. But I'll add itree to Coq CI soon.

PR #246, opened 2022-11-21
Add RuttFacts.v to _CoqProject.itree.

Chobbes:
I think this needs to be added to the _CoqProject.itree file in order to be able to import RuttFacts.v.


Lysxia:
Oops. Yeah, you need it if you're not using dune :)

PR #252, opened 2023-03-04
Fixing a couple of warnings

YaZko:
Explicit declaration of scopes and removing ignored implicit argument.


YaZko:
Is CI broken or am I missing something? This PR only fixes warning, and CI fails in relation to paco.

Lysxia:
There's a breaking change in paco master. Let me see if I can fix this.

YaZko:
Thanks @Lysxia !

PR #271, opened 2024-10-09
Updates for Paco v4.2.1

gilhur:
- Removed workaround code for a bug in `pclearbot`.
- The workaround is no longer necessary as the underlying bug has been fixed in Paco v4.2.1.


Lysxia:
Thanks Gil!

PR #272, opened 2025-01-30
 Updates for Paco v4.2.3

gilhur:
- Paco v4.2.3 fixes a bug in pclearbot, which now works better under binders
- This is an update for Paco v4.2.3


Lysxia:
Thanks!

